PR #7, opened 2017-05-09
User cert management

### LINKED ISSUE #8: Download or Generate user's certificate under "settings"
Users who have already logged in, need to view, download his cert files. If there's no cert for the current user, users can generate one.

typhoonzero:
Fix #8


wangkuiyi:
I think we need an issue describing the purpose and motivation of this PR.

PR #123, opened 2017-06-02
change restclient

gongweibao:


go/paddlecloud/restclient.go
```diff
@@ -14,7 +14,15 @@ import (
 // HTTPOK is ok status of http api call
 const HTTPOK = "200 OK"
 
-var httpTransport = &http.Transport{}
+type RestClient struct {
```
typhoonzero: What are the benefits of adding a struct? We can not re-use the connection to the server if we need to create a new `RestClient` every time.

gongweibao: 有道理。fix了。

gongweibao: 优势在于可以多线程并发。

go/paddlecloud/restclient.go
```diff
@@ -14,7 +14,15 @@ import (
...
+}
+
+// NewRestClient returns a new RestClient struct.
+func NewRestClient() *RestClient {
+	client := http.Client{Transport: &http.Transport{}}
+	return &RestClient{client: &client}
+}
 
 func makeRequest(uri string, method string, body io.Reader,
 	contentType string, query map[string]string,
```
gongweibao: 之所以改成`string`，是因为`map[string][]string`才能表达，而这样的结构体用起来太复杂，不如让用户在外边自己`encode`

go/paddlecloud/restclient.go
```diff
@@ -8,22 +8,31 @@ import (
...
+	"net/url"
 	"os"
 )
 
-// HTTPOK is ok status of http api call
+// HTTPOK is ok status of http api call.
 const HTTPOK = "200 OK"
 
-var httpTransport = &http.Transport{}
+type RestClient struct {
```
typhoonzero: A struct with only one member seems not necessary. If we want to implement multi-threading download processes, we can use a more simple function like:
```go
func GetChunk(targetURL string, query url.Values, nThreads int) (*http.Response, error) {
    req, err := makeRequest()
    // use a thread pool of http.client to do multi-threading
}
```

Then, we can call this functions by:
```go
resp := restclient.GetChunk()
```
instead of

```go
client := NewRestClient()
resp := client.GetChunk()
```

go/paddlecloud/restclient.go
typhoonzero: Cool.

go/paddlecloud/restclient.go
gongweibao: httpclient的确是线程安全的。Done。

PR #144, opened 2017-06-12
add testcase

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/156
Add test case of `ls` `chunkmeta` `cp`.

### DIFF
--- a/.tools/gen_config.sh
+++ b/.tools/gen_config.sh
@@ -1,0 +1,10 @@
+#/bin/bash
+mkdir -p ~/.paddle
+cat > ~/.paddle/config << EOF
+datacenters:
+- name: datacenter1
+  username: gongweibao@baidu.com
+  password: T123
+  endpoint: http://127.0.0.1:8080
+current-datacenter: datacenter1
+EOF
--- a/.tools/gen_config.sh
+++ b/.tools/gen_config.sh
@@ -3,7 +3,7 @@ mkdir -p ~/.paddle
 cat > ~/.paddle/config << EOF
 datacenters:
 - name: datacenter1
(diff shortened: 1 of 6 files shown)


.tools/gen_config.sh
```diff
@@ -0,0 +1,10 @@
+#/bin/bash
+mkdir -p ~/.paddle
+cat > ~/.paddle/config << EOF
+datacenters:
+- name: datacenter1
+  username: gongweibao@baidu.com
```
typhoonzero: Do not use actual usernames.

gongweibao: Done

.tools/gen_config.sh
```diff
@@ -0,0 +1,10 @@
+#/bin/bash
+mkdir -p ~/.paddle
```
helinwang: Maybe instead of generating this file with a script (unit test breaks without running this script), you can check in this file into the code base, maybe under `testdata`:https://github.com/golang/go/blob/master/src/net/testdata/hosts . And add argument to override the default location `~/.paddle` to the test file.

gongweibao:
@typhoonzero 修改了`config`全局变量，避免了对`~/.paddle/config`的依赖。所以关闭掉这个PR，重新在 @typhoonzero  的基础上提交一个。

.tools/gen_config.sh
gongweibao: `config`是一个全局变量，这个参数不太好覆盖。如果不用这个全局变量，需要改动的地方太多。这个能否后续再想想如何更简单的解决这问题？

PR #152, opened 2017-06-15
Add utils test

typhoonzero:
This is part work of https://github.com/PaddlePaddle/cloud/issues/11

### DIFF
--- a/go/utils/config_test.go
+++ b/go/utils/config_test.go
@@ -1,36 +1,49 @@
 package utils
 
 import (
+	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
+	"strconv"
 	"testing"
 )
 
-func fakeServer() *http.Server {
+func fakeServer() (*http.Server, int) {
 	http.HandleFunc("/api-token-auth/", func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("{\"token\": \"testtokenvalue\"}"))
 	})
 	http.HandleFunc("/fake-api/", func(w http.ResponseWriter, r *http.Request) {
 		w.Write([]byte("fakeresult"))
 	})
-	srv := &http.Server{Addr: ":8980"}
+	listener, err := net.Listen("tcp", ":0")
+	if err != nil {
+		panic(err)
+	}
+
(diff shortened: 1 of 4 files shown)


go/utils/restclient.go
```diff
@@ -17,8 +17,10 @@ import (
 // HTTPOK is ok status of http api call.
 const HTTPOK = "200 OK"
 
-var HttpClient = &http.Client{Transport: &http.Transport{}}
+// HTTPClient is a global Http client obj conains one persist connection.
+var HTTPClient = &http.Client{Transport: &http.Transport{}}
```
helinwang: Just curious, `http.DefaultClient` seems a reasonable starting point, what is the benefit of using `var HTTPClient = &http.Client{Transport: &http.Transport{}}`?

go/utils/config_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"net"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strconv"
+	"testing"
+)
+
+func fakeServer() (*http.Server, int) {
+	http.HandleFunc("/api-token-auth/", func(w http.ResponseWriter, r *http.Request) {
```
helinwang: Using `httptest` maybe more convenient: https://golang.org/pkg/net/http/httptest/#example_Server

go/utils/restclient.go
typhoonzero: Good point! I intended to make sure that the client can reuse the TCP connection for HTTP calls. I found this:

> A Client is an HTTP client. It's zero value (DefaultClient) is a usable client that uses DefaultTransport.

The Client's Transport typically has internal state (cached TCP connections), so Clients should be reused instead of created as needed. Clients are safe for concurrent use by multiple goroutines.

So `http.DefaultClient` seems to do the exact same thing!

go/utils/config_test.go
typhoonzero: Seem that `httptest` is a good choice for testing HTTP servers(server request handlers), but these test cases are test client calls. A "fake" server must be started with partial function to test all the client routines.

But It's a nice package! Will make use of it somewhere else!

helinwang: I see, thanks for explaining!

Issue #145, opened 2017-06-12
When job pod fails getting log returns error


typhoonzero:
![fde7fb61363e17b4a8e0a789a](https://user-images.githubusercontent.com/13348433/27031251-1ec923a6-4fa2-11e7-8daf-e21efbd72b6a.jpg)

PR #153, opened 2017-06-15
Fix log command error

typhoonzero:
Fix https://github.com/PaddlePaddle/cloud/issues/145

### DIFF
--- a/go/paddlecloud/logs.go
+++ b/go/paddlecloud/logs.go
@@ -53,7 +53,8 @@ func (p *LogsCommand) Execute(_ context.Context, f *flag.FlagSet, _ ...interface
 
 	respBody, err := utils.GetCall(utils.Config.ActiveConfig.Endpoint+"/api/v1/logs", queryMap)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "call paddle cloud error %v", err)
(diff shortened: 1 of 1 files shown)


go/paddlecloud/logs.go
```diff
@@ -53,7 +53,8 @@ func (p *LogsCommand) Execute(_ context.Context, f *flag.FlagSet, _ ...interface
 
 	respBody, err := utils.GetCall(utils.Config.ActiveConfig.Endpoint+"/api/v1/logs", queryMap)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "call paddle cloud error %v", err)
+		fmt.Fprintf(os.Stderr, "call paddle cloud error %v, %v", err, respBody)
```
helinwang: Maybe it's easier for viewing to print `string` instead of `[]byte`, I could be wrong...

typhoonzero: Printing `string` or `[]byte` can both use `%s`. Newline can also contain in a `string` which will affect the log printing

helinwang: I see, thanks! Did not know about `%s` can be used on `[]byte`.

typhoonzero: Yes, I did. both `%v` and `%s` works for `[]byte`

PR #204, opened 2017-07-07
Add cloud pre-commit.

gongweibao:
According https://github.com/PaddlePaddle/Paddle/pull/2643

### DIFF
--- a/.tools/check_style.sh
+++ b/.tools/check_style.sh
@@ -13,7 +13,7 @@ export PATH=/usr/bin:$PATH
 pre-commit install
 pre-commit  --version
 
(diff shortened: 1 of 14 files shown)


.tools/check_style.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+function abort(){
+    echo "Your change doesn't follow PaddlePaddle's code style." 1>&2
```
typhoonzero: PaddlePaddle => PaddleCloud

gongweibao: Done.

PR #240, opened 2017-07-24
Fix Chinese charaset error

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/230


k8s/mysql/paddlecloud.sql
```diff
@@ -0,0 +1 @@
+CREATE DATABASE IF NOT EXISTS paddlecloud DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_general_ci;
```
typhoonzero: Can you put this in the kubernetes yaml file as "command" instead of using a self image repo?

gongweibao: Done

Issue #256, opened 2017-07-27
Sometimes `paddlecloud logs` is delayed


helinwang:
I can see the logs maybe 1min after the job completed, was expecting to see almost realtime logs.
Just curious why logs are very delayed?

Update: the first jobs have delayed log, the second job's logs looks fine.

typhoonzero:
docker collect logs from stdout and since it is buffered the output is not "real time", will add `stdbuf -oL` to the start up script.

PR #259, opened 2017-07-28
Fix docker log buffer

typhoonzero:
Fix https://github.com/PaddlePaddle/cloud/issues/256


typhoonzero:
Will update in the next beta release!

PR #286, opened 2017-08-09
Add demo of understand_sentiment on paddlecloud

wanghaoshuang:
Related: #147


typhoonzero:
Oh, another thing, you can try to add the fault tolerant support reader for each demo and test them out.

PR #287, opened 2017-08-09
Fix ls public dataset bug

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/281

### DIFF
--- a/go/filemanager/pfsmodules/command.go
+++ b/go/filemanager/pfsmodules/command.go
@@ -49,8 +49,28 @@ func checkUser(pathStr string, user string) error {
 	return nil
 }
 
-// ValidatePfsPath returns whether a path is a pfspath.
-func ValidatePfsPath(paths []string, userName string) error {
+func isPublic(pathStr string) bool {
+	pathStr = path.Clean(strings.TrimSpace(pathStr))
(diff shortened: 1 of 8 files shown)


go/filemanager/pfsmodules/command.go
```diff
@@ -49,8 +49,28 @@ func checkUser(pathStr string, user string) error {
...
+	if len(a) >= 4 && a[3] == "public" {
+		return true
+	}
+
+	return false
+}
+
+func checkPublic(cmdName string) error {
+	switch cmdName {
+	case "ls", "stat":
```
helinwang: This limits the CLI to do only `ls` and `stat`, but does the backend still allow cmd like `rm` and `cp` to function? If so, user could easily patch here and run commands like `rm` to the public dataset.

gongweibao: 不会的。用户任务运行的时候，容器里边其自己的目录是读写挂载的，`public`目录是只读挂载的。
理论上说，`pfsserver`的容器只读挂载`public`，我这里不需要判断，Ceph自己处理。不过为了出错信息可读另外是参考一下这个: https://github.com/PaddlePaddle/cloud/issues/281#issuecomment-321186934，所以加上了。

PR #301, opened 2017-08-12
Fix ls public data bugs

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/281

### DIFF
--- a/k8s/pfs_deployment.yaml
+++ b/k8s/pfs_deployment.yaml
@@ -2,6 +2,7 @@ apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   name: pfsserver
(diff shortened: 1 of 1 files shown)


k8s/pfs_deployment.yaml
```diff
@@ -10,22 +11,34 @@ spec:
         app: pfsserver
     spec:
       volumes:
+      - name: public-data-storage
+        cephfs:
+          monitors: ["ip:6789"]
```
typhoonzero: Maybe add one line of comment to notify users to modify the ip?

gongweibao: Done.

PR #385, opened 2017-10-12
[WIP]Autoscaling Controller

typhoonzero:
Fix https://github.com/PaddlePaddle/cloud/issues/384

This is under development, I created this PR to show the differences.


typhoonzero:
Changes made:
- Moved API definations to `go/api`. Use `v1.ResourceRequirements` from `client-go` as Resource defination.
- Add jobparser.go to parse `TrainingJob` resource to corresponding `Job` and `ReplicaSet`.
- Move autoscaler to `go/controller/autoscaler`. Using `v1.ResourceRequirements` is precise but anoying, should refine this.

TODO:

- Complete jobparser.
- Sync status of thirdpartyresource `TrainingJob`.

go/controller/autoscaler/cluster.go
```diff
@@ -147,7 +147,7 @@ func (c *K8sCluster) SyncResource() error {
 		}
 		for resname, q := range item.Status.Allocatable {
 			if resname == v1.ResourceCPU {
-				totalCPU += float64(q.Value()) + float64(q.MilliValue())/1000.0
```
helinwang: Go constant is very special:

> Go is a statically typed language that does not permit operations that mix numeric types. You can't add a float64 to an int, or even an int32 to an int. Yet it is legal to write 1e6*time.Second or math.Exp(1) or even 1<<('\t'+2.0). In Go, constants, unlike variables, behave pretty much like regular numbers. This post explains why that is and what it means.

So it's more idiomatic to simple use `1000` here, don't worry about accidentally mess up anything, the Go compiler is very strict about type, so if anything could go wrong, it won't compile.

Please see: https://blog.golang.org/constants

go/controller/autoscaler/autoscaler.go
```diff
@@ -62,7 +63,6 @@ func NewAutoscaler(cluster Cluster, options ...func(*Autoscaler)) *Autoscaler {
 		cluster: cluster,
 		ticker:  time.NewTicker(defaultLoopDur),
 		jobs:    make(map[string]job),
-		mutex:   &sync.Mutex{},
```
helinwang: Don't need to initialize `sync.Mutex`, but need to initialize `*sync.Mutex`. People typically use `sync.Mutex`.

go/controller/autoscaler/autoscaler.go
```diff
@@ -52,8 +52,9 @@ func (j job) Fulfillment() float64 {
 type Autoscaler struct {
 	ticker  *time.Ticker
 	cluster Cluster
-	jobs    map[string]job
-	mutex   *sync.Mutex
+
+	mutex sync.Mutex
```
helinwang: It's a convention to define mutex on top of the vars that needs to be protected.

go/controller/autoscaler/autoscaler.go
```diff
@@ -182,6 +182,10 @@ func (a *Autoscaler) Monitor() {
 		select {
 		case <-a.ticker.C:
 		}
+		// TODO(helin): if we handle job add / delete by
```
helinwang: If `mutex.Lock` and `mutex.Unlock` spread in the code, it's possible to cause deadlock if not possible. In this case maybe is more suitable to use channel.

go/controller/autoscaler/autoscaler.go
```diff
@@ -181,6 +178,21 @@ func (a *Autoscaler) Monitor() {
 	for {
```
helinwang: Everything runs in this goroutine, no more data race, don't need mutex anymore.

go/controller/autoscaler/autoscaler.go
```diff
@@ -81,10 +81,9 @@ func (j jobs) Less(a int, b int) bool {
 	if scoreA == scoreB {
 		resA := j[a].Config.Spec.Trainer.Resources
 		resB := j[b].Config.Spec.Trainer.Resources
-		// FIXME: use Quantity type, should refine these code.
 		resALimitsGPU := resA.Limits[paddlejob.GPUResourceName]
 		resBLimitsGPU := resB.Limits[paddlejob.GPUResourceName]
-		if resALimitsGPU.Cmp(resALimitsGPU) == 0 {
+		if resALimitsGPU.Cmp(resBLimitsGPU) == 0 {
```
helinwang: It's hard to find bug without test :)

typhoonzero: Will add later.

helinwang: @typhoonzero Already have test here, I found this line after fixing the build of the test.

go/controller/autoscaler/autoscaler.go
```diff
@@ -31,7 +31,7 @@ type Cluster interface {
 }
 
 type job struct {
-	Config      *paddlejob.TrainingJob
+	Config      paddlejob.TrainingJob
```
typhoonzero: `paddlejob.TrainingJob` can be a large struct. We don't want to copy around this struct.

helinwang:
```go
var j api.TrainingJob
fmt.Println(unsafe.Sizeof(j)) // 472
```
With 1MB of memory, we can have around 2200 `api.TrainingJob`, maybe it's not that big?

Here is a related guide: [pass values](https://github.com/golang/go/wiki/CodeReviewComments#pass-values), a `string` mentioned in the link is 16 bytes, so I think this is not a strong case for "pass values". Anyway, something to consider.

go/controller/autoscaler/cluster.go
```diff
@@ -30,9 +30,6 @@ type K8sCluster struct {
 // NewK8sCluster create a new instance of K8sCluster.
 func NewK8sCluster(clientset *kubernetes.Clientset) *K8sCluster {
 	return &K8sCluster{
-		NodeCount: 0,
```
helinwang: It's more idiomatic not to initialize values as default values, since the program already do so.

go/autoscaler/k8s/cluster.go
```diff
@@ -12,20 +26,20 @@ import (
...
+// Cluster resprensents a Kubernetes cluster.
+type Cluster struct {
 	clientset *kubernetes.Clientset
 }
 
-// NewK8sCluster create a new instance of K8sCluster.
-func NewK8sCluster(clientset *kubernetes.Clientset) *K8sCluster {
-	return &K8sCluster{
+// NewCluster create a new instance of K8sCluster.
+func NewCluster(clientset *kubernetes.Clientset) *Cluster {
```
helinwang: Putting `K8sCluster` into `k8s` package, so it can be renamed to `Cluster`. From outside of the package, it will be referred as `k8s.Cluster`.

go/autoscaler/k8s/controller.go
```diff
@@ -48,21 +48,10 @@ type Controller struct {
...
-		return nil, err
-	}
-	client, err := rest.RESTClientFor(config)
-	if err != nil {
-		return nil, err
-	}
-	// TODO: init autoscaler with correct arguments.
-	cluster := autoscaler.NewK8sCluster(clientset)
-	as := autoscaler.NewAutoscaler(cluster)
+func NewController(c *rest.RESTClient, cs *kubernetes.Clientset, as *autoscaler.Autoscaler) (*Controller, error) {
```
helinwang: "Dependency injection": pass in the dependencies, rather than creating dependency. This can simplify testing.

go/autoscaler/autoscaler.go
```diff
@@ -82,8 +95,8 @@ type Autoscaler struct {
 	eventCh chan event
 }
 
-// NewAutoscaler creates a new Autoscaler.
-func NewAutoscaler(cluster Cluster, options ...func(*Autoscaler)) *Autoscaler {
+// New creates a new Autoscaler.
+func New(cluster Cluster, options ...func(*Autoscaler)) *Autoscaler {
```
helinwang: Rename to `New`, can be referred from outside of the package as `autoscaler.New`, rather than `autoscaler. NewAutoscaler`.

go/controller/autoscaler/cluster.go
```diff
@@ -34,138 +24,29 @@ func NewK8sCluster(clientset *kubernetes.Clientset) *K8sCluster {
...
-	return *trainerJob.Spec.Parallelism
-}
-
-// TODO(helin): refine scale algorithm. E.g., if we have two jobs, a
-// needs 3 more GPU, b needs 3 more GPU, and there are 4 free GPUs, a
-// and b should both get 2 GPUs, but according to the current logic, a
-// will get 3 GPUs and b will get 1 GPU.
-
-// Scale one job if there's enough resource.
-func (c *K8sCluster) Scale(job paddlejob.TrainingJob) error {
```
helinwang: Moved Scale into package `autoscaler`.

go/autoscaler/autoscaler.go
```diff
@@ -11,12 +11,15 @@
...
    See the License for the specific language governing permissions and
 	 limitations under the License. */
+
 package autoscaler
 
 import (
 	"sort"
 	"time"
 
+	// TODO(typhoonzero): this package still depends on k8s API, try to remove this.
```
helinwang: 确实，我理解依赖k8s的原因是我们的autoscaler没有自己的任务相关的结构体（现在用的是"github.com/PaddlePaddle/cloud/go/api"和"k8s.io/client-go/pkg/apis/batch/v1"里面的结构体），如果要写自己的结构体，好处是autoscaler这个包很稳定，不随着k8s的变化而变化。坏处是会多一步k8s和自己的结构体转换的步骤。现在因为autoscaler也不支持k8s以外的，可能直接用k8s的结构体会方便点。

go/autoscaler/autoscaler.go
```diff
@@ -259,13 +273,13 @@ func scaleAllDryRun(jobs []job, r ClusterResource) map[string]int {
 func (a *Autoscaler) scaleAll(diff map[string]int) {
 	for name := range diff {
 		if diff[name] != 0 {
-			log.Infoln("Scaling job %s, diff: %d.", name, diff[name])
+			log.Infof("Scaling job %s, diff: %d.", name, diff[name])
 			target := *a.jobs[name].TrainerJob.Spec.Parallelism + int32(diff[name])
 			*a.jobs[name].TrainerJob.Spec.Parallelism = target
-			*a.jobs[name].TrainerJob.Spec.Completions = target
+			// *a.jobs[name].TrainerJob.Spec.Completions = target
```
helinwang: 这个我当时没测，好奇为啥Completions不需要设呢？

go/autoscaler/autoscaler.go
typhoonzero: 嗯嗯。而且里面的Quantity很好用。后续可以想一些方法迭代下。

go/autoscaler/autoscaler.go
typhoonzero: See:

go/controller/cluster.go
```diff
@@ -50,82 +50,23 @@ func (c Cluster) GetTrainerJob(job *paddlejob.TrainingJob) (*batchv1.Job, error)
...
-		}
-		for podLimitName, podLimitValue := range podLimits {
-			if value, ok := limits[podLimitName]; !ok {
-				limits[podLimitName] = *podLimitValue.Copy()
-			} else {
-				value.Add(podLimitValue)
-				limits[podLimitName] = value
-			}
+		for _, container := range pod.Spec.Containers {
+			AddResourceList(&reqs, container.Resources.Requests)
```
typhoonzero: I tried to reduce the lines by using `AddResourceList`.

Removed getting resource from `InitContainers` because currently paddle job trainer seems do not use initcontainer.

go/controller/cluster.go
```diff
@@ -136,65 +77,41 @@ func (c *Cluster) SyncResource() (res autoscaler.ClusterResource, err error) {
...
+		AddResourceList(&allocatable, node.Status.Allocatable)
+	}
 
-		for key, item := range limits {
-			a := allLimits[key]
-			a.Add(item)
-			allLimits[key] = a
-		}
+	// get non-terminated pods from all namespaces all nodes.
+	// FIXME(typhoonzero): scan all pods is not a efficient way.
```
typhoonzero: We need to get pod resource requests from all the nodes. I simplified this by get all the pods in one API call.

typhoonzero: If there's too much pods, this may take a lot of memory though.

go/autoscaler/autoscaler.go
```diff
@@ -38,9 +38,9 @@ type ClusterResource struct {
 	GPULimit   int
 	GPUTotal   int
 
-	CPURequestKilo float64
-	CPULimitKilo   float64
-	CPUTotalKilo   float64
+	CPURequestMilli int64
```
typhoonzero: K8s CPU request can be at least `0.1` core which means `100m` (100 milli CPU), change kilo to milli

docker/k8s_tools.py
```diff
@@ -27,7 +27,8 @@ def wait_pods_running(label_selector, desired):
     print "label selector: %s, desired: %s" % (label_selector, desired)
     while True:
         count = count_pods_by_phase(label_selector, 'Running')
-        if count == int(desired):
+        # NOTE: pods may be scaled.
```
typhoonzero: Pods may be scaled.

go/autoscaler/autoscaler_internal_test.go
```diff
@@ -0,0 +1,543 @@
...
+	j.Config.Spec.Trainer.Resources.Requests = make(v1.ResourceList)
+	j.Config.Spec.Trainer.Resources.Requests["cpu"] = q1
+	j.Config.Spec.Trainer.Resources.Requests["memory"] = q100M
+	j.TrainerJob.Spec.Parallelism = makePtr(1)
+	assert.Equal(t, 1, scaleDryRun(&r, j, 0))
+}
+
+func TestScaleDryRunNoMoreCPU(t *testing.T) {
+	r := ClusterResource{
+		CPULimitMilli:     1000,
```
typhoonzero: 1000 milli("1000m") means 1 cpu core, "100m" means 0.1 cpu core.

k8s/controller/autoscale_job/trainer.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
       - name: trainer
         image: paddlepaddle/paddlecloud-job
-        imagePullPolicy: Always
+        imagePullPolicy: Never
```
helinwang: Curious why it's "Never"?

helinwang:
I saw "github.com/PaddlePaddle/cloud/go/cfs" being added, but currently not used and partially implemented. It contains https://github.com/sakeven/RbTree/blob/master/rbtree.go , maybe the best way is to use it as a dependency, or fork it.

If we need to add a scheduler, I think the most important thing to do now is to define a interface for the scheduler, and have a naive implementation for the interface. Later we can switch to CFS if we need. Currently the bottle neck is not the scheduler, given that the functionality is still not complete:
1. Controller can not submit jobs.
1. Controller does not register

go/controller/utils.go
```diff
@@ -18,16 +18,16 @@ import (
 	"k8s.io/client-go/pkg/api/v1"
 )
 
-// AddResourceList add another v1.ResourceList to first's inner quantity.
-// v1.ResourceList is equal to map[string]Quantity
-func AddResourceList(a *v1.ResourceList, b v1.ResourceList) {
+// AddResourceList add another v1.ResourceList to first's inner
+// quantity.  v1.ResourceList is equal to map[string]Quantity
+func AddResourceList(a v1.ResourceList, b v1.ResourceList) {
```
helinwang: `a` is of type `map`, don't need to pass as pointer: https://github.com/golang/go/wiki/CodeReviewComments#copying

I understand from the C programming language's perspective
`AddResourceList(&reqs, container.Resources.Requests)` is more clear than `AddResourceList(reqs, container.Resources.Requests)` about `reqs` has been modified. I think that's fine in Go, since `req` is a `map`.

go/controller/cluster.go
helinwang: @typhoonzero from the article we are going to use the controller in a general cluster, so could be pods other than Paddle pods.

go/controller/cluster.go
helinwang: I think the memory limit is fine, 1MB can fit thousands of "medium-sized" data structures.

go/controller/cluster.go
```diff
@@ -73,16 +73,21 @@ func (c Cluster) JobRunning(job *paddlejob.TrainingJob) (bool, error) {
...
+// limits from all pods containers.
 func getPodsTotalRequestsAndLimits(podList *v1.PodList) (reqs v1.ResourceList, limits v1.ResourceList, err error) {
 	reqs, limits = v1.ResourceList{}, v1.ResourceList{}
 	for _, pod := range podList.Items {
 		for _, container := range pod.Spec.Containers {
 			AddResourceList(reqs, container.Resources.Requests)
 			AddResourceList(limits, container.Resources.Limits)
 		}
+
+		for _, container := range pod.Spec.InitContainers {
```
helinwang: from ChenXi's article we are going to use the controller in a general cluster, so could be pods other than Paddle pods.

k8s/controller/autoscale_job/trainer.yaml
typhoonzero: Oh, That's for my local running, will update.

typhoonzero:
@helinwang 

> I think If we need to add a scheduler, we need to first define an interface for the scheduler, have a naive implementation for the interface, and defer the decision about which kind of scheduler to implement for as long as possible (in this way we have more information to make the correct decision). Currently the bottleneck is not the scheduler, given that the functionalities are still not complete:

I intended to push some codes that may be useful in the future. Currently we do scale by scaning the pods one by one without consider the priority or faireness of the jobs. This may due to some "important" jobs may never get scaled. I'll add some design doc later to try to make clear of this.

I think I'll keep the choice of CFS here because we intend to let all job share resources equally(can have job weights). I'll leave the interface here and add a design doc.

go/controller/cluster.go
typhoonzero: You are right! Thanks!

helinwang:
@typhoonzero I see, thanks!

helinwang:
@typhoonzero I see, thanks! Maybe we can first deploy our current implementation on the cluster, and work on scheduler after we get more experience with real usage.

Issue #409, opened 2017-10-18
demo/fit_a_line is not working properly


putcn:
fit_a_line is not working properly even without fault tolerance.

putcn:
closed with fix #410

PR #410, opened 2017-10-18
fix demo fit_a_line

putcn:
fixing #409 
with original fc config, the default learning rate will lead to gradient explosion. fixed by adding learning_rate=1e-3

### DIFF
--- a/demo/fit_a_line/train_ft.py
+++ b/demo/fit_a_line/train_ft.py
@@ -14,7 +14,11 @@ def main():
 
     # network config
     x = paddle.layer.data(name='x', type=paddle.data_type.dense_vector(13))
-    y_predict = paddle.layer.fc(input=x, size=1, act=paddle.activation.Linear())
+    y_predict = paddle.layer.fc(
+        input=x, 
+        size=1, 
+        act=paddle.activation.Linear(), 
+        param_attr=paddle.attr.Param(learning_rate=1e-3))
     y = paddle.layer.data(name='y', type=paddle.data_type.dense_vector(1))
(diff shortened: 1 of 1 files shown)


demo/fit_a_line/train_ft.py
```diff
@@ -14,7 +14,11 @@ def main():
 
     # network config
     x = paddle.layer.data(name='x', type=paddle.data_type.dense_vector(13))
-    y_predict = paddle.layer.fc(input=x, size=1, act=paddle.activation.Linear())
+    y_predict = paddle.layer.fc(
+        input=x, 
+        size=1, 
+        act=paddle.activation.Linear(), 
+        param_attr=paddle.attr.Param(learning_rate=1e-3))
```
typhoonzero: `bias_attr` also need to set. Anyway, this is due to the incorrect convertion from old optimizer configuration to new optimizer configuration of paddle.

putcn: got it, thanks @typhoonzero. will check and update later.

helinwang: > this is due to the incorrect convertion from old optimizer configuration to new optimizer configuration of paddle.

@typhoonzero Can you create a issue with a link indicating the line number which have the problem? This could provide better documentation so that in case in the future we run into it again, we can look at the issue and know what happened. Thanks!

typhoonzero: @helinwang Sure! I added: https://github.com/PaddlePaddle/Paddle/issues/5005 .

PR #439, opened 2017-10-26
Replace logrus with log15, update glide files. Add retry limit when get trainer job errors.

helinwang:
Fixes: https://github.com/PaddlePaddle/cloud/issues/437


helinwang:
@typhoonzero Yes, log15 is quite flexible, let's add a flag if we feel the logs lines are too long.

PR #452, opened 2017-10-30
Set debug level to master's log.

gongweibao:
Set debug level to master's log.

### DIFF
--- a/docker/paddle_k8s
+++ b/docker/paddle_k8s
@@ -23,7 +23,8 @@ start_new_pserver() {
 start_master() {
   stdbuf -oL /usr/bin/master \
   -port=8080 \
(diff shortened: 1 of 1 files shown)


docker/paddle_k8s
```diff
@@ -23,7 +23,8 @@ start_new_pserver() {
 start_master() {
   stdbuf -oL /usr/bin/master \
   -port=8080 \
-  -endpoints=http://127.0.0.1:2379
+  -endpoints=http://127.0.0.1:2379 \
+  -log-level=debug
```
typhoonzero: This may cause too much log if the user can't turn it off, can you please change it to configurable, please?

Issue #512, opened 2017-12-04
Ending slash in `~/.paddle/config` cause  an `404 not found server` error!


gongweibao:
If  `endpoint` ends with `/`:
```
- name: mulan
  username: work_beijing@163.com
  password: 1
  endpoint: http://yxxx:8880/
```
it will cause an error:
```
./pcloud -v=10 ls /pfs/mulan/home/xxx.com/
previous token not found, fetching a new one...
/pfs/mulan/home/work_beijing@163.com/ :
  error:get token error:server error: 404 Not Found
```

PR #513, opened 2017-12-04
Fix ending slash bugs.

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/512

### DIFF
--- a/go/utils/config/config.go
+++ b/go/utils/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"io/ioutil"
 	"path/filepath"
+	"regexp"
 
 	"github.com/PaddlePaddle/cloud/go/utils/pathutil"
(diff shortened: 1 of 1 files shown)


go/utils/config/config.go
```diff
@@ -48,6 +49,12 @@ func ParseConfig(configFile string) *SubmitConfig {
 			glog.Errorf("load config %s error: %v\n", configFile, yamlErr)
 			return nil
 		}
+
+		var re = regexp.MustCompile(`(/|\\)*$`)
```
typhoonzero: Regex is slow, why not simply remove the trailing slash. By the way, can you add a verify of the endpoint string to ensure it is an URL?

gongweibao: 1. The user's input is diverse and only a few strings are checked.
2. Done!

PR #540, opened 2018-01-03
support v1 split train.list

typhoonzero:
Fix https://github.com/PaddlePaddle/cloud/issues/539


typhoonzero:
I'll appreciate if you can do that, the submit part is also needed.

PR #544, opened 2018-01-04
add the process of building autoscaler controller

m3ngyang:
Fix https://github.com/PaddlePaddle/cloud/issues/541

### DIFF
--- a/doc/autoscale/README.md
+++ b/doc/autoscale/README.md
@@ -140,6 +140,9 @@ The training job controller can be started by the cluster
 administrator with command: `kubectl create -f
 training_job_controller.yaml`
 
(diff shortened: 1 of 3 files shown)


doc/autoscale/README.md
```diff
@@ -140,6 +140,9 @@ The training job controller can be started by the cluster
 administrator with command: `kubectl create -f
 training_job_controller.yaml`
 
+You can use `go/cmd/autoscaler/Dockerfile` to build a new controller image 
+or downloading an existing one.
```
typhoonzero: Maybe add a link to dockerhub?

m3ngyang: yes, we should push an image to official paddlepaddle namespace on dockerhub, instead of using `registry.baidu.com`

PR #565, opened 2018-01-27
implement registration and listening of crd trainingjob

m3ngyang:
I have implemented a custom resource definition for paddle training job. The changes in this PR include:
1. `go/pkg/controller`, responsible for the registration and listening for CRD TrainingJob, and creating a job tracker for each training job to manage its whole lifecycle
2. `go/pkg/client`, auto generated codes for CRD client and informer
3. `go/hack`, use our own boilerplate file as default rights comment in generated codes
4. `go/cmd/operator`, a new binary component for CRD, which will be combined with the autoscaler in the near feature

About 1st change, I leave a interface for @qizheng09 to implement the job tracker/controller.


go/pkg/controller/controller.go
```diff
@@ -0,0 +1,248 @@
...
+	// ApiCli is the extension kubernetes clientset
+	ApiCli apiextensionsclient.Interface
+	// PaddleCli is a clientset for our own API group
+	PaddleCli paddleclientset.Interface
+
+	trainingjobLister paddlelisters.TrainingJobLister
+	trainingjobSynced cache.InformerSynced
+
+	// TODO jobtracker keep track of every training job
+	jobtracker map[string]interface{}
```
m3ngyang: the interface will be replaced with specific type for managing training job.

go/pkg/controller/controller.go
```diff
@@ -0,0 +1,248 @@
...
+		}
+
+		return err
+	}
+
+	_, ok := c.jobtracker[key]
+	if !ok {
+		glog.Infof("create a new job tracker, key: '%s'", key)
+		glog.Infof("received job: %+v", job)
+		// TODO create a tracker for the job, just record its name here
```
m3ngyang: TODO create a tracker for the job here

typhoonzero:
1. This PR is too large to review. Or may take too long. We should keep each PR small so that we can merge it fast.
1. Why we need a new directory `pkg`? If we should have one it's better to rename the top level `go` to `pkg`.
1. Better to not have `hack` directory, if it's some hacking code that messing up the code, it should be very carefully reviewed.

Anyway, can you please split this PR to several small PRs so that I can review.

m3ngyang:
I will split the pr into three parts:
1. reorganize the directory, rename the `go` dir to `pkg` and move the dirs under current `pkg` dir
2. supplement the generated codes for crd
3. add the implementation of management of crd
Is this reasonable? @typhoonzero

typhoonzero:
@m3ngyang Sure!

Issue #586, opened 2018-02-04
.travis.yml refers to nonexisting bash script


wangkuiyi:
`.travis.yml` refers to `.tools/gen_config.sh`

https://github.com/PaddlePaddle/cloud/blob/48b4261342213445eeded108612c3916d4231894/.travis.yml#L18

which doesn't seem exist in https://github.com/PaddlePaddle/cloud/tree/develop/.tools

wangkuiyi:
It seems that in [the PR](https://github.com/PaddlePaddle/cloud/pull/148) by @gongweibao , the file .tools/gen_config.sh was moved into [`go/.tools/gen_config.sh`](https://github.com/PaddlePaddle/cloud/tree/develop/go/.tools)

gongweibao:
Yep.

wangkuiyi:
https://github.com/PaddlePaddle/cloud/blob/develop/go/.tools/gen_config.sh

What is the purpose of this shell script? Is it necessary?

PR #603, opened 2018-02-06
Cleanup useless `gen_config.sh`.

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/586


typhoonzero:
Compile fails
```
# github.com/PaddlePaddle/cloud/go/paddlectl
paddlectl/submit.go:25: undefined: config.SumitConfig
```

And, need update the unit test if default config is not generated.

go/filemanager/pfsmodules/upload.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 // Config is global config object for pfs commandline
-var Config = config.ParseDefaultConfig()
+var Config *config.SubmitConfig
```
typhoonzero: Why `Config` global is defined in `upload.go`, should it be placed in a common file?

gongweibao: Done

PR #613, opened 2018-02-09
fix: style check does not work

### LINKED ISSUE #609: CI doesn't exit when meets a lint error!
```
RESULT=$?; if [ $RESULT -eq 0 ]; then true; else false; fi;
```

m3ngyang:
fix #609

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,8 @@ matrix:
       script:
         - rm -f .copyright.hook && wget https://raw.githubusercontent.com/PaddlePaddle/Paddle/develop/.copyright.hook
         - |
-          bash .tools/check_style.sh
-          RESULT=$?; if [ $RESULT -eq 0 ]; then true; else exit 1 ; fi;
+          bash -x .tools/check_style.sh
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,8 +10,8 @@ matrix:
       script:
         - rm -f .copyright.hook && wget https://raw.githubusercontent.com/PaddlePaddle/Paddle/develop/.copyright.hook
         - |
-          bash .tools/check_style.sh
-          RESULT=$?; if [ $RESULT -eq 0 ]; then true; else exit 1 ; fi;
+          bash -x .tools/check_style.sh
+          RESULT=$?; if [ $RESULT -eq 0 ]; then echo Passed ; else echo Failed >&2 ; fi;
```
typhoonzero: Can you take alook at https://github.com/PaddlePaddle/cloud/pull/614? I think it should be better than this?

m3ngyang: yes, my pr does not fix it well.

typhoonzero: OK, closing now, thanks!

Issue #616, opened 2018-02-09
Unify the docker images build method


typhoonzero:
We now have the following docker images need to be built to run on the cluster:

- EDL controller
- Cloud server
- PFS server
- Cloud job runtime image

The `Dockerfile`s and build method is different, need a unified way or a script to build them all.

typhoonzero:
Or, we can put all component in one Docker image and run several pods using the same image.

PR #619, opened 2018-02-11
Unify dockerfile

gongweibao:
Fix https://github.com/PaddlePaddle/cloud/issues/616


go/Dockerfile
```diff
@@ -1,3 +1,3 @@
 FROM ubuntu:16.04
-ADD ./cmd/pfsserver/pfsserver /pfsserver/
-RUN mkdir /pfsserver/log
+ADD ./cmd/pfsserver/pfsserver /usr/bin/
```
typhoonzero: Can we move all `Dockerfiles` under `docker/`?

PR #620, opened 2018-02-11
move glide files to the top directory

### LINKED ISSUE #615: [WIP]Move glide.yaml and glide.lock to the top directory
This could be helpful that we can run `glide install -v` directly under the top-level directory, then the `vendor` directory is not created inside `go` folder which could be annoying.

m3ngyang:
fix #615

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ matrix:
       go: 1.8.x
       sudo: required
       install:
+        - sudo apt-get install -y mercurial
         - go get -u github.com/golang/lint/golint
         - curl https://glide.sh/get | bash
         - sudo pip install pre-commit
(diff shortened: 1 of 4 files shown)


typhoonzero:
@Yancey1989 how to build and test go code under `doc` folder?

m3ngyang:
@typhoonzero problem solved, but we cannot get `gonum.org/v1/plot` directly.
We need to set a mirror for the repo when using `glide install` locally.
`glide mirror set https://gonum.org/v1/plot https://github.com/gonum/plot`

typhoonzero:
> but we cannot get gonum.org/v1/plot directly.

This may not be a problem for Go developers in China.

m3ngyang:
This is not caused by the internet connection. There is a discussion [here](https://github.com/gonum/gonum/issues/260) and I use glide mirror to walk around the problem.

typhoonzero:
I see, thanks!

PR #626, opened 2018-02-28
adding usage_en doc

putcn:
related: #594


putcn:
thanks for the review @Yancey1989, updated

PR #624, opened 2018-02-27
change api version of crd to v1

m3ngyang:
This pr is part of #565, includes two commits:
1. change api version of trainingjob to v1 and fix errors in scripts after the change
2. generate tools codes for crd

PR #628, opened 2018-03-01
commit TrainingJobUpdater struct

qizheng09:
This pr commit the `TrainingJobUpdater` struct

### DIFF
--- a/go/updater/trainingJobUpdater.go
+++ b/go/updater/trainingJobUpdater.go
@@ -1,0 +1,81 @@
+package updater
+
+import (
+	padv1 "github.com/PaddlePaddle/cloud/go/apis/paddlepaddle/v1"
+	trainingJobClient "github.com/PaddlePaddle/cloud/go/client/clientset/versioned"
+	log "github.com/golang/glog"
+	"k8s.io/client-go/kubernetes"
+	"time"
+)
+
+const (
+	retryTimes            = 5
+	convertedTimerTicker  = 10 * time.Second
+	confirmResourceTicker = 5 * time.Second
+)
+
+type trainingJobEventType string
+
+const (
+	trainingJobEventDelete trainingJobEventType = "Delete"
+	trainingJobEventModify trainingJobEventType = "Modify"
+)
+
+type trainingJobEvent struct {
+	// pet is the TrainingJobEventType of TrainingJob
+	pet trainingJobEventType
+	// The job transfer the information fo job
+	job *padv1.TrainingJob
+}
+
+// TrainingJobUpdater is to manager a specific TrainingJob
+type TrainingJobUpdater struct {
+	// Job is the job the TrainingJob manager.
+	job *padv1.TrainingJob
+
+	// kubeCli is standard kubernetes client.
+	kubeCli kubernetes.Interface
+
+	// TrainingJobClient is the client of TrainingJob.
+	trainingJobClient trainingJobClient.Interface
+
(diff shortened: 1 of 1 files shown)


go/updater/trainingJobUpdater.go
```diff
@@ -0,0 +1,81 @@
...
+	// kubeCli is standard kubernetes client.
+	kubeCli kubernetes.Interface
+
+	// TrainingJobClient is the client of TrainingJob.
+	trainingJobClient trainingJobClient.Interface
+
+	// Status is the status in memory, update when TrainingJob status changed and update the CRD resource status.
+	status padv1.TrainingJobStatus
+
+	// EventCh is the channel received by Controller, include Modify and Delete.
```
typhoonzero: > is the channel received by Controller

Do you mean "this channel is shared with the controller to receive TrainingJob events"

go/updater/trainingJobUpdater.go
```diff
@@ -0,0 +1,81 @@
...
+	// When trainingJobEvent is Delete it will delete all resources
+	// The maximum is 1000.
```
typhoonzero: What maximum here? The capacity or time?

go/updater/trainingJobUpdater.go
```diff
@@ -0,0 +1,81 @@
...
+// TrainingJobUpdater is to manager a specific TrainingJob
+type TrainingJobUpdater struct {
+	// Job is the job the TrainingJob manager.
+	job *padv1.TrainingJob
+
+	// kubeCli is standard kubernetes client.
+	kubeCli kubernetes.Interface
+
+	// TrainingJobClient is the client of TrainingJob.
+	trainingJobClient trainingJobClient.Interface
```
typhoonzero: What is the difference between `kubeCli` and `trainingJobClient`? Shall we use pointer here?

go/updater/trainingJobUpdater.go
qizheng09: The comments should be changed to "received from Controller". Controller use `EventCh` to notify updater.

go/updater/trainingJobUpdater.go
qizheng09: The maximum is the capacity of the `EventCh`.  For a `TrainingJob`,  1000 may enough.

go/updater/trainingJobUpdater.go
qizheng09: `kubeCli` only have the functions of k8s native resources. We need to use the `trainingJobClient` which @m3ngyang committed in  [#624](https://github.com/PaddlePaddle/cloud/pull/624).

m3ngyang: > Shall we use pointer here?

Interface is the better choice. In golang, interface is a spec that defines the behavior of a given object and Kubernetes also adopts Interface-based programming.

typhoonzero: @m3ngyang I see, thanks!

go/updater/trainingJobUpdater.go
typhoonzero: Or maybe like: "EventCh receives event from the controller"?

go/updater/trainingJobUpdater.go
typhoonzero: Okay.

go/updater/trainingJobUpdater.go
qizheng09: Okay~

PR #633, opened 2018-03-05
Fix fluid start entry

typhoonzero:
Job start command should not have command line arguments added by the hidden script. Use env instead.

### DIFF
--- a/docker/paddle_k8s
+++ b/docker/paddle_k8s
@@ -65,10 +65,12 @@ start_fluid_process() {
   task_index=""
 
   stdbuf -oL python /root/k8s_tools.py wait_pods_running ${pserver_label} ${PSERVERS}
-  stdbuf -oL python /root/k8s_tools.py wait_pods_running ${trainer_label} ${TRAINERS}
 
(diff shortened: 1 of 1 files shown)


docker/paddle_k8s
```diff
@@ -65,20 +65,23 @@ start_fluid_process() {
...
     task_index=$(python /root/k8s_tools.py fetch_id ${trainer_label})
   else
     task_index=$(python /root/k8s_tools.py fetch_id ${pserver_label})
   fi
+  
+  export PADDLE_INIT_TRAINER_ID=${task_index}
 
-  stdbuf -oL sh -c "${ENTRY} --ps_hosts=${ps_hosts} --trainer_hosts=${trainer_hosts} \
-      --task_index=${task_index}"
+  stdbuf -oL sh -c "${ENTRY}"
```
gongweibao: 1. Should add `task_index` environment.

typhoonzero: `PADDLE_INIT_TRAINER_ID` does the job.

typhoonzero: > how can the users print all env variables if not using the command line

using python os.envs

